PR #54, opened 2017-10-08
Add table styled line numbers

### LINKED ISSUE #52: Add table styled line numbers
See https://github.com/gohugoio/hugo/issues/3915

bep:
Fixes #52

### DIFF
--- a/cmd/chroma/main.go
+++ b/cmd/chroma/main.go
@@ -47,6 +47,7 @@ var (
 	htmlInlineStyleFlag    = kingpin.Flag("html-inline-styles", "Output HTML with inline styles (no classes).").Bool()
 	htmlTabWidthFlag       = kingpin.Flag("html-tab-width", "Set the HTML tab width.").Default("8").Int()
 	htmlLinesFlag          = kingpin.Flag("html-lines", "Include line numbers in output.").Bool()
+	htmlLinesTableFlag     = kingpin.Flag("html-lines-table", "Split line numbers and code in a HTML table").Bool()
 	htmlLinesStyleFlag     = kingpin.Flag("html-lines-style", "Style for line numbers.").String()
 	htmlHighlightFlag      = kingpin.Flag("html-highlight", "Highlight these lines.").PlaceHolder("N[:M][,...]").String()
 	htmlHighlightStyleFlag = kingpin.Flag("html-highlight-style", "Style used for highlighting lines.").String()
@@ -139,6 +140,9 @@ command, for Go.
 		}
 		if *htmlLinesFlag {
 			options = append(options, html.WithLineNumbers())
+		}
+		if *htmlLinesTableFlag {
+			options = append(options, html.LineNumbersInTable())
 		}
 		if len(*htmlHighlightFlag) > 0 {
 			ranges := [][2]int{}
--- a/formatters/html/html.go
+++ b/formatters/html/html.go
@@ -32,6 +32,14 @@ func WithLineNumbers() Option {
 	}
 }
 
+// LineNumbersInTable will, when combined with WithLineNumbers, separate the line numbers
+// and code in table td's, which make them copy-and-paste friendly.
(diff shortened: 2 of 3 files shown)


bep:
Notes:

I have added `highlighttable´ class to the table. This matches Pygments, but I have not added any styles to it. Pygments also has a class on both the td's (linenos, code), but I dropped that.

alecthomas:
I think the output is not quite right. Have you checked it?

I see this with no tables:
image

And this with tables:
image

bep:
> Have you checked it?

Yes. But it depends on what you mean by "not quite right". If you mean "not same", you are right.

alecthomas:
The version with tables has the line numbers seemingly (ie. visibly) off by one. Notice that LANGS is almost on line 2 in the second example.

 My guess is it's something to do with vertical padding/margin.

bep:
Yes, I can make the padding the same, but could you shed some light on my bullet 2) above. I.e. how do you handle CSS diffs from "invoking the CLI with different params". Ideally I want to have a mix of table and inline styled line number types and have one stylesheet to match them both.

Right now it is:

```css
/* LineNumbers */ .chroma .ln { color: #7f7f7f; margin-right: 0.4em; padding: 0 0.4em 0 0.4em;; display: block; }
```

Which I notice also contains an error, which I will fix.

The same with inline line numbers:

```bash
/* LineNumbers */ .chroma .ln { color: #7f7f7f; margin-right: 0.4em; padding: 0 0.4em 0 0.4em; }
```

alecthomas:
> We should probably handle this better so people can switch between these with the same stylesheet... Suggestions welcome.

Ah, sorry, it wasn't clear to me what "these" was in this context.

I haven't looked at the details, but the class might need to be special-cased for the case when lines are in a table. This should be possible in `styleToCSS()`, maybe...

bep:
>but the class might need to be special-cased for the case

There is already a special case, but please read and understand my last post.

alecthomas:
Why would anyone want a mix of table and inline line-number styles simultaneously? Unless there's a compelling reason, I would not want to support that.

bep:
Define "simultaneously" ... So:

* From Pygments there are 1 stylesheet for (what they call) `inline` and `table` styled line numbers.
* Hugo supports both Chroma and Pygments.
* Hugo supports setting the "table vs inline" style on a global and per code block level
* If I want to create a Hugo theme I would want to supply some Chroma stylesheet, but it would be up to the end user if he prefers  `inline` or `table` styled line numbers
* In many cases it would make sense to use only one and stick with it, but if I want to change my mind, I want to just "toggle a switch" in a config file.

I can come up with more if you need it.

alecthomas:
> Ideally I want to have a mix of table and inline styled line number types and have one stylesheet to match them both.

This statement sounds to me like you wanted to have code blocks with inline numbers and code blocks with table numbers in the same page at the same time. That is what I was referring to.

From my perspective, as long as the HTML and CSS output by Chroma result in a decent-looking rendered output, I'm happy. Anything beyond that I don't really care about that much, but if you want to come up with a clean way to support both with the same CSS, go for it. If you'd prefer to output different CSS for each, based on the formatter options, then do that.

bep:
>This statement sounds to me like you wanted to have code blocks with inline numbers and code blocks with table numbers in the same page at the same time. That is what I was referring to.

This is not me wanting things. And the above sounds like something "someone wants". I will have a look at a way to fix this.

bep:
No table: 

image

With table: 

image

I have taken another stab at this. A couple of things to notice:

1. There is a tab diff in the above screenshots. The tab width used is 4, but in the "non table" variant this looks to include the line numbers, which does not make much sense to me. 
2. I looked at the Pygments "way of doing this", but that looked messy, and the classes used are not included in the Pygments stylesheets, so I just made up a couple of new CSS classes.

Let me know of any objections ...

alecthomas:
Looks great, thanks @bep!

PR #117, opened 2018-02-10
go: Fix line comment at end of input

jgrahn:
Chroma's go lexer did not correctly recognise single line comments if they were at the end of the input without a trailing newline. This PR corrects that issue.

While having input that ends without a final newline is not normally the case with source code files, it often occurs when using chroma as a library to highlight code snippets displayed in other tools. In particular, this is the case with [Hugo](https://gohugo.io/). Adding a newline is not a valid workaround, as that can cause dead space to appear below the code snippet.

Given the following input:

    var x = 1 // Comment 1
    var y = 2 // Comment 2<no-new-line>

...output **before** the PR was:
screenshot from 2018-02-10 22-54-56

...and output **after** the PR is:
screenshot from 2018-02-10 22-54-31

### DIFF
--- a/lexers/go.go
+++ b/lexers/go.go
@@ -19,7 +19,7 @@ var Go = Register(MustNewLexer(
 			{`\n`, Text, nil},
 			{`\s+`, Text, nil},
 			{`\\\n`, Text, nil},
-			{`//(.*?)\n`, CommentSingle, nil},
+			{`//(.*?)$`, CommentSingle, nil},
 			{`/(\\\n)?[*](.|\n)*?[*](\\\n)?/`, CommentMultiline, nil},
 			{`(import|package)\b`, KeywordNamespace, nil},
 			{`(var|func|struct|map|chan|type|interface|const)\b`, KeywordDeclaration, nil},


alecthomas:
There is an EnsureNL lexer configuration option to control this behavior.

jgrahn:
Ah, then it sounds like I should make another PR to the downstream maintainers and have them set that option.

Nevertheless, accepting this PR would make chroma interpret go sources the same way the go compiler does.

alecthomas:
Sorry, I'm talking about the go lexer configuration.

jgrahn:
Alright, then that's probably a better solution.

alecthomas:
> Adding a newline is not a valid workaround, as that can cause dead space to appear below the code snippet.

Are you sure this is true? If it's the case that `line\nline\n` adds an empty line at the end of the snippet, I'd think that is a bug.

jgrahn:
To be precise, what I am sure of is that the following input:
```
{{< highlight go >}}
// This is a comment
{{< / highlight >}}
```
results in this output in Hugo:
screenshot from 2018-02-11 01-16-10

while the following:
```
{{< highlight go >}}
// This is a comment (next line has a blank space)
 
{{< / highlight >}}
```
results in this output:
screenshot from 2018-02-11 01-19-46

However, and here is the crucial part, adding any number of newlines at the end without any other trailing characters (like the space above) yields the first example. So it seems Hugo trims any trailing `\n` characters _before_ sending the string to chroma, and presumably expects them to still be trimmed in the colourised output. (I haven't dived into that code and verified this so far -- I'm already much deeper into this rabbit whole than I was planning to be. :slightly_smiling_face:)

Whether to label this a bug in Hugo or not I'm not sure. It's almost a matter of opinion. I have no axe to grind, though, as long as the end result looks good.

alecthomas:
This is an inherited "feature" from Pygments, for better or worse. You can read the history of it [here](https://github.com/alecthomas/chroma/issues/42). Basically, most of the Pygments lexers contain this same assumption, so fixing them ad-hoc is not really the right solution.

`EnsureNL` is the correct fix for now. If you want to modify this PR to do that then I'll merge it, otherwise I'll close this out.

alecthomas:
This is actually a very annoying feature of Pygments. Your solution is the correct one IMO, but it needs to be applied to all lexers.

jgrahn:
I now had a quick look at a bunch of the other lexers to get a sense of how much work it would be to fix all of them. Turns out they're actually not consistent in how they handle this case. Some require explicit newline, some are okay with end-of-line.

There could of course be languages where inputs not ending with a final newline character are invalid or have a different meaning, explaining such differences, but since most code snippets aren't valid programs anyway, perhaps the most sensible thing is to always treat any such case as if there were a newline following it (i.e. accept end-of-line in its place) for the purposes of highlighting.

alecthomas:
Yes, I think fixing it would be a quite a bit of work :(

PR #179, opened 2018-09-22
Pass Token around by value instead of pointer

eloff:
Hi,

Here's a basic attempt at changing to Token, and changing use of nil token for (Token{}, false).

Sometimes this gets a little clumsy, such as in delegate.go. Probably the code could be improved with a less literal translation, but that's all I had the imagination for at this hour on a Friday night.

The lexer benchmark improves slightly:
Benchmark-20                1000           1237538 ns/op           89612 B/op       1599 allocs/op

vs master:
Benchmark-20                1000           1242267 ns/op           78404 B/op       1723 allocs/op

I feel instinctively that it should be a bigger difference. So I think there's some work to do to figure out where the other allocations are coming from.

What do you think of this direction so far?

### DIFF
--- a/cmd/chroma/main.go
+++ b/cmd/chroma/main.go
@@ -259,7 +259,11 @@ func format(w io.Writer, style *chroma.Style, it chroma.Iterator) {
 
 func check(filename string, it chroma.Iterator) {
 	line, col := 1, 0
-	for token := it(); token != nil; token = it() {
+	for {
+		token, ok := it()
+		if !ok {
+			break
+		}
 		if token.Type == chroma.Error {
 			fmt.Printf("%s:%d:%d %q\n", filename, line, col, token.String())
 		}
--- a/coalesce.go
+++ b/coalesce.go
@@ -6,30 +6,38 @@ func Coalesce(lexer Lexer) Lexer { return &coalescer{lexer} }
 type coalescer struct{ Lexer }
 
 func (d *coalescer) Tokenise(options *TokeniseOptions, text string) (Iterator, error) {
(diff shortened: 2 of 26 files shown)


alecthomas:
Hmmm. I appreciate the effort, but I think with the minimal improvement in time/garbage overhead, coupled with the increase in code ugliness, I'm -1 on this.

If the code remained fairly similar to what it is now, maybe via having a sentinel "EOF" value that could be substituted for the current nil check? That would keep the code clean but still allow the minimal improvement in garbage.

eloff:
Keep in mind that the benchmark is probably not representative. The real-life gain could be smaller or larger. The sample code for the benchmark is very small, so the number of tokens is low compared to the other overhead during parsing. So I think we shouldn't trust the benchmark as a good data point. I would like to come up with a more representative one. I'd also like to know where the other allocations are happening.

That having been said, I agree completely with what you're saying. An "nil" or EOF token value would clean up the code a lot, in fact it's not really any different than the previous code at that point, just without the allocations.

The obvious value is the zero token. Token{}. Is that unambiguous?

alecthomas:
> The obvious value is the zero token. Token{}. Is that unambiguous?

It should be okay (IIRC there's no TokenType of 0)... it would be good to add an `EOF TokenType = 0`.

eloff:
Ok, I'll do that, and I'll see about getting better benchmark info.

alecthomas:
Any progress?

eloff:
None, I'm on vacation, it will likely have to wait until late next week
when I'm back.

PR #215, opened 2019-01-15
Fix: recognise end comment in TradingView lexer

Jos512:
I noticed that the TradingView lexer does not recognise a comment when a code snippet ends with a comment.

The cause for that is a requirement in the lexer that there should be a newline (`\n`) after a comment. That applies to all comments, except the last one in a code highlight.

This pull request:

- Changes the lexer so that the newline requirement is dropped.
- Adds an example comment as the last line in the testdata.
- Updates the expected outcome of the testdata.

## Before
(blue = comment)

before-example

## After this pull request

after-example

----

Let me know if you have any feedback or preferred changes.

Thanks for considering this pull request. :slightly_smiling_face:

### DIFF
--- a/lexers/t/tradingview.go
+++ b/lexers/t/tradingview.go
@@ -17,7 +17,7 @@ var TradingView = internal.Register(MustNewLexer(
 	Rules{
 		"root": {
 			{`[^\S\n]+|\n|[()]`, Text, nil},
-			{`(//.*?)(\n)`, ByGroups(CommentSingle, Text), nil},
+			{`//.*?$`, CommentSingle, nil},
 			{`>=|<=|==|!=|>|<|\?|-|\+|\*|\/|%|\[|\]`, Operator, nil},
 			{`[:,.]`, Punctuation, nil},
 			{`=`, KeywordPseudo, nil},
--- a/lexers/testdata/tradingview.actual
+++ b/lexers/testdata/tradingview.actual
@@ -10,3 +10,5 @@ plot(series=emaVal, style=circles, offset=2, linewidth=3)
 bgcolor(color=close > open ? orange :
      close != high[1] ? purple :
      na, transp=80)
+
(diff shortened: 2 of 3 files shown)


alecthomas:
Hi @Jos512 . I appreciate these pull requests, but this is actually a deeper problem. There's a lexer option called "EnsureNL" which should, ideally, add a newline to lexers like this one that require newlines in order to match. The problem is that it also used to add a complete new line to highlighted blocks. This may be fixed now, but I'm not sure.

Jos512:
Thanks for the feedback Alec. If I enable `EnsureNL` for the TradingView lexer, it does recognise the last line comment. And, when adding a border to the `chroma` class, doesn't add a new line it seems:



I see a similar thing with the C# lexer, when that one has `EnsureNL` enabled:



Is this what you meant with the 'complete new line'-behaviour? 

If I look at the HTML output, I see *without* `EnsureNL` (so `master` behaviour):

~~~html
<span class="c1">// Comment example</span>
<span class="o">/</span><span class="o">/</span> Include the other strategy code here</pre>
~~~

And with `EnsureNL` enabled:

~~~html
<span class="c1">// Comment example</span>
<span class="c1">// Include the other strategy code here</span>
</pre>
~~~

So the `</pre>` end tag does move, and isn't on the same line anymore as the last highlighted line. But the browser (FF latest and Chrome latest) doesn't render a complete new line.

alecthomas:
Interesting. I did recently [fix a bug](881a441774f9d707d3b7852025b7f2149a556182) with too many newlines being emitted, so maybe it's fixed...

Jos512:
Do you welcome a pull request that uses `EnsureNL` with the TradingView and C# lexer, or would you prefer some further testing and experimenting?

alecthomas:
That would be great, thank you!

PR #275, opened 2019-09-01
Add italics support for truecolour terminals

raylee:
This PR is a bit of a request for feedback. I made the smallest change to support my use-case
for emitting italics, namely limiting it to terminals which claim to support truecolour.

It's possible this change should be replicated in tty_indexed.go as well. (Perhaps some indexed color terminal emulators support italics?) I'm happy to add that change if that's preferred, but I don't have an example emulator to point to where that's true.

Below is a screenshot with the change in iTerm2 and Source Code Pro, showing `go doc ioutil | chroma -f terminal16m -s algol`

Screen Shot 2019-09-01 at 12 21 36 PM


raylee:
Sweet, I've updated the PR. Thanks!

PR #423, opened 2020-11-11
Update supported languages table

pabluk:
This PR update the `table.py` script allowing to write the content directly into `README.md` to avoid manual actions and maybe could be useful to automate the release steps.
The table of supported languages was update using the script.

I tried to keep the script as simple and flat as possible, using only Python 3.x built-in functions. I though about the idea to implemented with a `--write-in-place` option but I don't know if it worth it.

In any case, feel free to request any changes you think could be appropriate, provide any feedback, or even just drop this PR because I know that adding more code implies more maintenance. Thanks for your time!


alecthomas (approved):
I like it!

PR #465, opened 2021-04-05
fix: synthesisable entries are ignored when cloned

zhangyuannie:
For example, if `chroma` is invoked on the command line, `LineNumbers`, `LineHighlight` et al are ignored.

### DIFF
--- a/style.go
+++ b/style.go
@@ -265,11 +265,11 @@ func (s *Style) Get(ttype TokenType) StyleEntry {
 
 func (s *Style) get(ttype TokenType) StyleEntry {
 	out := s.entries[ttype]
-	if out.IsZero() && s.synthesisable(ttype) {
-		out = s.synthesise(ttype)
-	}
 	if out.IsZero() && s.parent != nil {
 		return s.parent.get(ttype)
+	}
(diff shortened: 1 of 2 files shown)


zhangyuannie:
When we try to acquire a mutable builder from a style and then use it:

`styleBar = styleFoo.Builder().Build()`

The resulting style `styleBar` will not use `LineHighlight`, `LineNumbers` or any other synthesisable attributes from `styleFoo`.

**To Reproduce**

1. use a style that specify `LineNumbers`, we can modify the builtin monokai style by adding:

   ```
   chroma.LineHighlight:       "bg:#ffffff",
   chroma.LineNumbers:         "bg:#ffffff",
   ```

   to `styles/monokai.go`.

2. Run with cmd interface.

   ```js
   // test.js
   function inc(n) {
     return n + 1;
   }
   ```

   `chroma -s monokai --html --html-lines --html-highlight=1:2 test.js`

**Expected**: some lines are white:

expected

**Received**: the highlighted lines' style is synthesised.

received

**This PR fixes this issue.**

alecthomas:
Thanks for the clarification!

PR #468, opened 2021-04-10
Simplify the cmdlet ("NameBuiltin") rule

LeeHolmes:
This change simplifies the NameBuiltin pattern for cmdlets so that it highlights all cmdlets (i.e.: Invoke-Foobar) and not just ones built into PowerShell. It also places these as higher priority than language keywords, so you don't get odd formatting for Foreach-Object and similar.

image


LeeHolmes:
Current output in the playground:

image

PR #486, opened 2021-05-04
lexers: match known backup filename suffixes

scop:
Perhaps not the cleanest way to do this, but conveys the idea and results in a small diff.


scop:
Moved suffix matching to Match after all, maybe somewhat cleaner this way. Not sure if these should be attempted after all non-backup filename and alias filename matches instead of interleaved, but leaving this way for comments.

PR #490, opened 2021-05-05
Use Python 3 by default (finish the job)

mlpo:
Oups, `MimeTypes` were incomplete (follows #489)…


mlpo:
Sorry, history pollution due to redundant commits, I will redo a clean PR

Issue #492, opened 2021-05-05
Add FortranFixedLexer


ProgramFan:
I am trying to add FortranFixedLexer from pygments using the bundled scripts. The script reports the following warning and the missing functions are added as invalid code in the generated go source. What can I do?

```
warning: unsupported emitter function <function FortranFixedLexer._lex_fortran at 0x7f8d2e6941f0>
warning: unsupported emitter function <function FortranFixedLexer._lex_fortran at 0x7f8d2e6941f0>
```

Generated go file:

```go

package lexers

import (
    . "github.com/alecthomas/chroma" // nolint
    "github.com/alecthomas/chroma/lexers/internal"
)

// Fortranfixed lexer.
var Fortranfixed = internal.Register(MustNewLazyLexer(
    &Config{
        Name:      "FortranFixed",
        Aliases:   []string{ "fortranfixed",  },
        Filenames: []string{ "*.f", "*.F",  },
        MimeTypes: []string{  },
        NotMultiline: true,
        CaseInsensitive: true,
    },
    func() Rules {
        return Rules{
            "root": {
                { `[C*].*\n`, Comment, nil },
                { `#.*\n`, CommentPreproc, nil },
                { ` {0,4}!.*\n`, Comment, nil },
                { `(.{5})`, NameLabel, Push("cont-char") },
                { `.*\n`, Using(Fortran), nil },
            },
            "cont-char": {
                { ` `, Text, Push("code") },
                { `0`, Comment, Push("code") },
                { `.`, GenericStrong, Push("code") },
            },
            "code": {
                { `(.{66})(.*)(\n)`, ByGroups(?? <function FortranFixedLexer._lex_fortran at 0x7f8d2e6941f0> ??, Comment, Text), Push("root") },
                { `(.*)(\n)`, ByGroups(?? <function FortranFixedLexer._lex_fortran at 0x7f8d2e6941f0> ??, Text), Push("root") },
                Default(Push("root")),
            },
        }
    },
))
```

ProgramFan:
After reading the pygments implementation, I realized that the missing function is actually a line lexer of Fortran. Replacing it directly with `Using(Fortran)` will compile and work most of the time.

ProgramFan:
After some manual tweak, the FortranFixedLexer now works.

PR #498, opened 2021-05-06
Add FortranFixedLexer

ProgramFan:
This PR ports the FortranFixedLexer from pygments using the bundled tools and some manual tweak. It works on ScaLapack code so I think it generally shall work. It addresses #492

### DIFF
--- a/lexers/f/fortran_fixed.go
+++ b/lexers/f/fortran_fixed.go
@@ -1,0 +1,39 @@
+package f
+
+import (
+	. "github.com/alecthomas/chroma" // nolint
+	"github.com/alecthomas/chroma/lexers/internal"
+)
+
+// FortranFixed lexer.
+var FortranFixed = internal.Register(MustNewLazyLexer(
+	&Config{
+		Name:            "FortranFixed",
+		Aliases:         []string{"fortranfixed"},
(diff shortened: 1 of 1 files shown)


lexers/f/fortran_fixed.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/alecthomas/chroma/lexers/internal"
+)
+
+// FortranFixed lexer.
+var FortranFixed = internal.Register(MustNewLazyLexer(
+	&Config{
+		Name:            "FortranFixed",
+		Aliases:         []string{"fortranfixed"},
+		Filenames:       []string{"*.f", "*.F"},
+		MimeTypes:       []string{"text/x-fortran"},
```
alecthomas, line 14: I don't know what FortranFixed is, but in hindsight it's weird that `.f` doesn't match the fortran lexer?

ProgramFan:
Thanks for the merge. Fortran has two forms: the fixed form and the free
form, thus two lexers are needed. The.f extension is usually used for fixed
form sources and the FortranLexer only recognize free forms, so we need a
new FortranFixedLexer.

Alec Thomas ***@***.***>于2021年5月7日 周五06:47写道：

PR #522, opened 2021-06-10
Added uint8_t to c lexer.

ktprograms:
Copied the KeywordType rule from the c++ lexer (but removed wchar_t since it doesn't exist in c).

Unfortunately, I can't test it since I'm unfamiliar with go.


alecthomas (approved):

ktprograms:
Thanks! BTW how/when will these changes be available in Hugo (the `highlight` shortcode)?

PR #570, opened 2021-11-06
Add WitchHazel by Thea Flowers

stefanv:
https://witchhazel.thea.codes/

### DIFF
--- a/styles/witchhazel.go
+++ b/styles/witchhazel.go
@@ -48,5 +48,5 @@ var WitchHazel = Register(chroma.MustNewStyle("witchhazel", chroma.StyleEntries{
 	chroma.GenericInserted:   "#a6e22e",
 	chroma.GenericStrong:     "bold",
 	chroma.GenericSubheading: "#75715e",
-	chroma.Background:        "#433e56",
+	chroma.Background:        " bg:#433e56",
 }))
--- a/styles/witchhazel.go
+++ b/styles/witchhazel.go
@@ -1,0 +1,52 @@
+// Copyright 2018 Alethea Katherine Flowers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


stefanv:
I have tested the addition using `go test`, but I would like to do some more testing (and visually inspect the output) before removing draft status.

Please forgive my ignorance, but I am new at go and cannot figure out how to install the binary for testing (`go build` and `go install` are both quiet).

alecthomas:
go build will place it in the working directory

alecthomas:
However you should add some actual tests to really be sure. Follow the README in lexers.

stefanv:
`go build` does not produce anything as far as I can tell.  It also looks like none of the other styles have tests associated with them.

alecthomas:
Ah sorry, I didn't read the code - the tests are for the lexers. `go build` will only produce an executable for a `main` package. I would build the chroma executable and use that:

```
go run ./cmd/chroma -s witchhazel SOMEFILE
```

stefanv:
Thanks, @alecthomas, that's helpful.  This is where my inexperience with go is a problem; I'm sure it must be obvious, but when I run `go run ./cmd/chroma` I get:

```
go run: no packages loaded from ./cmd/chroma
```

Maybe I need some environment variable set, or a newer version of go? I have:

```
go version go1.17.3 linux/amd64
```

Feel free to just point me to a manual or website; I can teach myself if I have a rough sense of where to look.

alecthomas:
Sorry that is my bad. I forgot that I made `chroma` a separate Go module. You'll need to do:

```
cd ./cmd/chroma
go run . -s witchhazel SOMEFILE
```

stefanv:
This worked well, and I was able to confirm that the theme renders as expected:

alecthomas:
Awesome, thanks :)

PR #633, opened 2022-05-03
fix: parse all of `@foo.bar` as a Python decorator

mhmdanas:


lexers/embedded/python.xml
```diff
@@ -379,7 +379,7 @@
       </rule>
     </state>
     <state name="name">
-      <rule pattern="@[_\p{L}][_\p{L}\p{N}]*">
+      <rule pattern="@[_\p{L}][_\p{L}\p{N}]*(\s*\.\s*[_\p{L}][_\p{L}\p{N}]*)*">
```
mhmdanas, line 382:
```suggestion
      <rule pattern="@[_\p{L}][_\p{L}\p{N}]*(\.[_\p{L}][_\p{L}\p{N}]*)*">
```

Accounting for whitespace feels wrong to me, what do you think?

alecthomas (approved):

PR #680, opened 2022-10-05
Add .pylintrc to ini lexer

silverwind:
Reference: https://pylint.pycqa.org/en/latest/user_guide/usage/run.html#command-line-options


silverwind:
Another reference from linguist: https://github.com/github/linguist/blob/1fbaa41dc259695226bb4ffd70f43957df1b7fb6/lib/linguist/languages.yml#L2780

alecthomas (approved):

PR #696, opened 2022-11-02
Improve Kotlin lexer

pmwmedia:
I have added a heuristic for classifying object names (something like static classes) and enum names as class names. This change adapts the behavior of the Java and Scala lexers.


alecthomas (approved):
Nice, thanks!

PR #713, opened 2022-11-24
lexers: Add benchmark for lexers.Get

bep:
```
BenchmarkLexersGet/Known-10             112448803            10.57 ns/op           0 B/op           0 allocs/op
BenchmarkLexersGet/Unknown-10                286       4160740 ns/op           0 B/op           0 allocs/op
```


walles:
The benchmark shows 4ms for the unknown case.

@bep, what kind of performance would have made you feel it was not slow?

PR #732, opened 2023-01-18
Add default diff support

CosmicHorrorDev:
This adds support for the default diff syntax. It looks like this is missing from `pygments` too, so I'll have to open a PR with them tomorrow

I didn't know what token type was appropriate for the separator, so I went with `GenericStrong`, but feel free to suggest something else

## Demo

`default.diff`

```diff
1,2d0
< A
< A
4c2
< C
---
> F
5a4
> E
```

Generated with

`old.txt`

```text
A
A
B
C
D
```

`new.txt`

```text
B
F
D
E
```

```bash
$ diff old.txt new.txt > default.diff
```

### Results

```bash
chroma --format="tokens" default.diff
```

_Before_

```go
&Token{Text, "1,2d0\n< A\n< A\n4c2\n< C\n"}
&Token{GenericDeleted, "---\n"}
&Token{Text, "> F\n5a4\n> E\n"}
```

_After_

```go
&Token{GenericSubheading, "1,2d0\n"}
&Token{GenericDeleted, "< A\n< A\n"}
&Token{GenericSubheading, "4c2\n"}
&Token{GenericDeleted, "< C\n"}
&Token{GenericStrong, "---\n"}
&Token{GenericInserted, "> F\n"}
&Token{GenericSubheading, "5a4\n"}
&Token{GenericInserted, "> E\n"}
```


alecthomas (approved):
Nice, thanks!

PR #739, opened 2023-01-29
Rust: Include starting apostrophe in lifetime / loop label token

CosmicHorrorDev:
Previously the apostrophe that would start lifetimes and loop labels would be considered an `Operator` while the ident after would be considered a `NameAttribute` (or `NameBuiltin` for `_` or `static`). This change includes the apostrophe in the token instead

This did involve pulling out the `(static|_)` to make that match distinct. Let me know if you know of a better way to handle this

The test file with syntax highlighting for reference

```rust
struct Bar<'b>(&'b str);

struct Foo<'b, 'a: 'b>(&'a Bar<'b>);

fn foo<'some_name>(_: Bar<'_>, _: impl Iterator<Item = Bar<'static>> + 'static) {
    'im_1a_loop_label: loop {
        break 'im_a_loop_label;
    }

    // None of these should have lifetimes
    let _ = 'a';
    let _ = b'a';
    let _ = "
        'not_a_lifetime_or_label
    ";
}
```


alecthomas (approved):
If it works, LGTM. Thanks!

PR #788, opened 2023-05-02
lexers/d: Update keyword lists

CyberShadow:
Add missing keywords, remove non-keywords, move a few things around to where they make more sense to be.

References:
- https://dlang.org/spec/lex.html#keywords
- https://dlang.org/spec/lex.html#specialtokens
- https://dlang.org/phobos/object.html


CyberShadow:
There is a bug: in `static if`, `if` is parsed as `NameFunction` (but it should be just a keyword). I'm not sure how to fix that - I don't see anything special about `static` in the XML.

PR #837, opened 2023-08-24
Set priority for INI lexer over Inform 6

gandarez:
The file extension `.inf` is far more common as `INI` than `Inform 6` since both share the same file extension.

https://en.wikipedia.org/wiki/INF_file
https://www.ifwiki.org/Inform_6


alecthomas (approved):

gandarez:
@alecthomas will you merge it soon?

Issue #847, opened 2023-09-15
Support Justfile


SteveLauC:
**What problem does this feature solve?**

Gitea uses `chroma` for syntax highlight, It would be great if `chroma` would support [justfile](https://github.com/casey/just) so that [Gitea could benefit from this](https://github.com/go-gitea/gitea/issues/27086)

**What feature do you propose?**

Support syntax highlight for justfile

silverwind:
If it looks somewhat okay while highlighted as `Makefile`, it could be added as an alias to is as a temporary solution until full support is added.

silverwind:
<img width="716" alt="image" src="https://github.com/alecthomas/chroma/assets/115237/7762f97a-fecb-44eb-95df-0fecb33b01e0">

Looks okay imho.

SteveLauC:
Thanks for doing this!

PR #852, opened 2023-09-16
add justfile filename to Makefile lexer

silverwind:
Fixes https://github.com/alecthomas/chroma/issues/847.


silverwind:
Added title case variant as well. The filename is actually case-insensitive, but I think these two variants would be the most common.

PR #882, opened 2023-11-13
Add Hare lexer

mhmdanas:
TODO:
- [x] Clean up
- [x] Add test(s)


lexers/hare.go
```diff
@@ -0,0 +1,60 @@
+package lexers
```
alecthomas, line 1: Please convert this to an XML lexer, I'm no longer accepting Go-based lexers unless there's a good reason. Conversion is fairly mechanical, see other lexers in lexers/embedded for examples.

mhmdanas: Done.

Just a couple suggestions regarding XML lexer format errors: I had spent some time trying to debug why the embedded lexer wasn't working for any Hare code I tried:
```console
$ chroma --json x.ha 
[
  {"type":"Error","value":"//test\nlet x = 1;\n"}
]

```

Until I figured out that I was supposed to wrap the `<include>` elements in `<rule>` elements; it would be great if Chroma printed out a clear error for this case, with a specific line and character or range. After that, I tried putting two `<include>` elements in one rule, and I got another somewhat cryptic error (which was fixed by putting each `<include>` in a separate rule):
```console
$ chroma --json x.ha
chroma: error: Hare: failed to compile rules: embedded/hare.xml: duplicate mutator
```

This is better than before, but it would be great if the error was more specific and it pointed to a specific line and character or range as mentioned already.

lexers/embedded/hare.xml
```diff
@@ -0,0 +1,104 @@
...
+  <config>
+    <name>Hare</name>
+    <alias>hare</alias>
+    <filename>*.ha</filename>
+    <mime_type>text/x-hare</mime_type>
+  </config>
+  <rules>
+    <state name="whitespace">
+      <rule pattern="^use\s.+;">
+        <token type="CommentPreproc"/>
```
mhmdanas, line 11: I don't like how this is done, it makes the whole `use` statement a uniform and weird greenish color:
image

Compare to Helix, which highlights each part of the statement separately:
image

alecthomas: You can use `<bygroups>` to achieve this.

mhmdanas: Thanks, I decided to have `use` be a `KeywordNamespace` but not use a separate state or `bygroups` for the rest of the `use` statement, as I thought it was a lot of hassle for little benefit as it already looks fine with my change:
image

lexers/embedded/hare.xml
```diff
@@ -0,0 +1,101 @@
...
+      <rule pattern="[\s\n]+">
+        <token type="TextWhitespace"/>
+      </rule>
+      <rule pattern="@[a-z]+">
+        <token type="NameDecorator"/>
+      </rule>
+      <rule pattern="//.*\n">
+        <token type="CommentSingle"/>
+      </rule>
+    </state>
```
mhmdanas, line 19: I don't see a good reason to have separate `whitespace` and `statements` states, maybe they should just be folded into the `root` state?

lexers/testdata/hare.actual
```diff
@@ -0,0 +1,66 @@
+use bufio;
```
mhmdanas, line 1: It's probably a good idea to have some test data that's a bit simpler and also uses more language constructs (e.g. enums, switch/match statements).

lexers/embedded/hare.xml
```diff
@@ -0,0 +1,101 @@
...
+  <config>
+    <name>Hare</name>
+    <alias>hare</alias>
+    <filename>*.ha</filename>
+    <mime_type>text/x-hare</mime_type>
+  </config>
+  <rules>
+    <state name="whitespace">
+      <rule pattern="[\s\n]+">
+        <token type="TextWhitespace"/>
```
mhmdanas, line 11: I'm not quite sure if this should be `"Text"` instead?

lexers/embedded/hare.xml
```diff
@@ -0,0 +1,101 @@
...
+      </rule>
+      <rule pattern="@[a-z]+">
+        <token type="NameDecorator"/>
```
mhmdanas, line 14: As far as I know attributes (as Hare calls `@test` and friends) are compiler directives, so should I use `CommentPreproc` here instead?

alecthomas: I think NamDecorator is probably more like what other languages use.

lexers/embedded/hare.xml
```diff
@@ -0,0 +1,101 @@
...
+      <rule pattern="0x[0-9a-fA-F]+[pP][+-]?\d+(f32|f64)">
+        <token type="LiteralNumberFloat"/>
+      </rule>
+      <rule pattern="0x[0-9a-fA-F]+(z|[iu](8|16|32|64)?)?">
+        <token type="LiteralNumberHex"/>
+      </rule>
+      <rule pattern="0o[0-7]+(z|[iu](8|16|32|64)?)?">
+        <token type="LiteralNumberOct"/>
+      </rule>
+      <rule pattern="0b[01]+(z[iu](8|16|32|64)?)?">
```
YerinAlexey, line 49: This doesn't highlight `0b1010u16` properly, as well as incorrectly catching `0b1010zu8`.

```suggestion
      <rule pattern="0b[01]+(z|[iu](8|16|32|64)?)?">
```

lexers/embedded/hare.xml
```diff
@@ -0,0 +1,101 @@
...
+      <rule pattern="[~!%^&amp;*+=|?:&lt;&gt;/-]|[ai]s\b|\.\.\.">
+        <token type="Operator"/>
+      </rule>
+      <rule pattern="[()\[\],.{};]">
+        <token type="Punctuation"/>
+      </rule>
+      <rule pattern="use\b">
+        <token type="KeywordNamespace"/>
+      </rule>
+      <rule pattern="(_|align|break|const|continue|else|enum|export|for|if|return|size|static|struct|offset|union|fn|free|assert|abort|alloc|let|len|def|type|match|switch|case|append|delete|insert|defer|yield|vastart|vaarg|vaend)\b">
```
YerinAlexey, line 64: It would be nice to handle both size as a type and size as a builtin.

I came up with this solution (before types rule, and with `size` added to types as a fallback):

```xml
<rule pattern="(size)([\s\n]*)(\()">
  <bygroups>
    <token type="Keyword" />
    <token type="TextWhitespace" />
    <token type="Punctuation" />
  </bygroups>
</rule>
```

alecthomas:
LGTM!

PR #918, opened 2024-01-22
Add support for Rego syntax

jamietanna:
Adapted heavily from several of the lexers in this repo, and
https://github.com/tsandall/vim-rego/blob/b47bf95996b72d630464b44ca567f2945bd37ac9/syntax/rego.vim

### DIFF
--- a/lexers/embedded/rego.xml
+++ b/lexers/embedded/rego.xml
@@ -1,0 +1,88 @@
+<lexer>
+  <config>
+    <name>Rego</name>
+    <alias>rego</alias>
+    <filename>*.rego</filename>
+  </config>
+  <rules>
+    <state name="root">
+      <rule pattern="(package|import|as|not|with|default|else|some|in|if|contains)\b">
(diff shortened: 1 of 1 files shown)


jamietanna:
Argh sorry @alecthomas  it looks like this hasn't actually worked OK - it looks like some of the text doesn't get recognised correctly.

2024-01-22-212019_1445x673_scrot

Any hints as what best to do here?

Happy to revert this PR until we've fixed it!

alecthomas:
No worries

I'd add a test and iterate until it passes

jamietanna:
Interesting, thanks - will give that a go tomorrow!

alecthomas:
FYI you can also use the chroma CLI to manually test your lexer: `chroma --lexer <file.xml> <input>`

jamietanna:
Thanks I had actually been doing that 🫣 but I changed themes to improve the output, but should have gone back to the default so I knew I'd not finished it off

PR #935, opened 2024-02-26
lexers/cue: support definitions and dollars in field names

myitcv:
'$' is valid in a bare field name in CUE. It is not special. It happens to be used by convention at the start of field names, but that is about it.

Definitions start with '#'.

Add a "section" of tests that cover the various types of field. There are no errors in new "tests", whereas before (i.e. without the change to the CUE lexer) there would have been.


alecthomas (approved):

myitcv:
Thanks for the fast merge of the PR. 

Do you have any indication as to when you will be cutting a new version please?

PR #966, opened 2024-05-22
Fix stack popping logic in Raku lexer

donaldh:
This fixes an array index bounds error.

### DIFF
--- a/lexers/raku.go
+++ b/lexers/raku.go
@@ -1366,11 +1366,11 @@ func popRule(rule ruleReplacingConfig) MutatorFunc {
 
 		if ok && len(stack) > 0 {
 			// Pop from stack
-			stack = stack[:len(stack)-1]
 			lastRule := stack[len(stack)-1]
 			lastRule.pushState = false
 			lastRule.popState = false
 			lastRule.pop = true
+			stack = stack[:len(stack)-1]
 			state.Set(stackName, stack)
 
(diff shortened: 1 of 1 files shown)


alecthomas:
This fails the tests for some reason.

donaldh:
It's been a while since this issue first arose but here is some quick info. When I run the latest hugo from brew on my site:
```
hugo v0.125.7+extended darwin/amd64 BuildDate=2024-05-08T14:46:24Z VendorInfo=brew

ERROR Could not render org: could not write output: runtime error: index out of range [-1]. Using unrendered content.
```
I will need to patch chroma to output file and line info to narrow down what triggers this.

donaldh:
Now that I have added filenames to the hugo logging, I am getting multiple errors and it's different files each run. That suggests it's racy and there's some shared state between goroutines. When I set `HUGO_NUMWORKERMULTIPLIER=1` to serialise the hugo build, I no longer see the error.

Apologies for the noise, it is an issue somewhere else, though I'm not sure if the bug is in chroma, go-org or hugo.

Looking back at the change I made, moving the pop operation was obviously the wrong thing to do. However an array bounds error will happen if the stack length == 1 when popRule() is called. Perhaps this should never happen, but maybe the condition should be `if ok && len(stack) > 1 {`

CIAvash:
Yeah, probably.

PR #1002, opened 2024-09-21
Update TypeScript lexer to allow nested generics

fredrare:
https://github.com/alecthomas/chroma/issues/425 shows that generics are not correctly identified in general, but they are rather being treated as JSX elements. I proposed a simple solution in the comments by adding a space between `<` and the word next to it, but I believe most people will either not find the solution or some of them will find it rather unappealing.

For this reason, I made the JSX rules recursive and added a `","` `Punctuation` token inside so that there can be a number of generics used, as well as allowing nested generics. While I am not really fond of this hack, given that generics are already treated as JSX elements, I think this is a fair and easy enough solution for most cases.

#### Before
Screenshot 2024-09-20 at 9 28 05 PM

#### With spacing solution
Screenshot 2024-09-20 at 9 30 13 PM

#### With recursive JSX and `","` `Punctuation` token
Screenshot 2024-09-20 at 9 55 11 PM


alecthomas:
Mind adding some JSX samples to the test? We should make sure that still works.

fredrare:
Sure!
I have just included a JSX example with usual attributes and syntax to the test (my bad for not including them before). It should work just fine with any old JSX elements. The only thing is that JSX now *also* allows for recursive elements, but the highlighter should render exactly the same output for any old valid JSX.

Screenshot 2024-09-21 at 5 58 42 AM

alecthomas:
Awesome, thanks!

alecthomas:
Thanks for fixing that, appreciate it.

PR #1006, opened 2024-10-02
Pager friendly terminal formatting

walles:
Let's say a pager, like [moar](https://github.com/walles/moar) (uses Chroma for syntax highlighting) or `less`, shows a line in the middle of a file.

Unless that line starts with the correct formatting for the line, the pager would have to scan the whole file from the start to get the coloring of this single line right.

Before this PR, lines were not guaranteed to start with formatting, but could sometimes rely on formatting from the preceding lines.

With this change in place, lines can now stand by themselves, and paging will work better on Chroma's output.

Missing formatting at the start of the line would happen when a token had a linefeed in the middle.

### DIFF
--- a/formatters/tty_indexed.go
+++ b/formatters/tty_indexed.go
@@ -1,7 +1,6 @@
 package formatters
 
 import (
-	"fmt"
 	"io"
 	"math"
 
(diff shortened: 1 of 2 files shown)


alecthomas (approved):
Nice, good idea!

PR #1129, opened 2025-08-24
Modelica language support alecthomas/chroma#1122

BasilioPV:
Add XML lexer and test files for Modelica language support


alecthomas:
Thanks!

PR #1195, opened 2026-01-13
Create a lexer for Markless

Shinmera:
Markless is an open document markup standard specified at

https://shirakumo.org/docs/markless

Reference implementations of full parsers, a test suite, sample documents, and more can be found on the same page.

This lexer implements a simpler version of Markless without maintaining the standard component and directive stack. It also does not distinguish every possible component type in Markless, as Pygments/Chroma's types lack appropriate markers.

Nevertheless the lexer is useful enough to properly mark up syntax constructs in Markless documents, especially aiding in distinguishing syntactical elements from actual textual content at a glance.

The lexer also does not properly handle nesting, as I don't think Pygments/Chroma is capable of that, at least as far as I could tell from the documentation and other parsers? If I'm wrong, please let me know how I can deal with, for example, the following:

```
# foo **//bar//**
```

Wherein *the entire line should be "GenericHeading"*, `# `,  `**`, and `//` should be "Keyword" and `bar` should be both "GenericStrong" and "GenericEmph".

Also please let me know if you'd rather have an XML version of the parser. I found the Go version easier to write, so that's what I did. However, I can also transcribe it into XML if necessary.

Thanks a lot for all your work on Chroma!


alecthomas:
Thanks!

PR #1197, opened 2026-01-20
Add `.env.*` pattern to bash lexer

silverwind:
Reference: https://vite.dev/guide/env-and-mode#env-files


lexers/embedded/bash.xml
```diff
@@ -13,6 +13,7 @@
     <filename>*.eclass</filename>
     <filename>.env</filename>
     <filename>*.env</filename>
+    <filename>.env.*</filename>
```
silverwind:
```suggestion
    <filename>.env.*</filename>
    <filename>*.env</filename>
```

PR #1203, opened 2026-01-26
Fix Readme table generation

JakobDev:
Currently the `chroma` binary needs to be in` PATH` for this script to work. This is a bit weird while developing. With this PR the script now runs Chroma directly from source.

### DIFF
--- a/table.py
+++ b/table.py
@@ -5,7 +5,7 @@ from subprocess import check_output
 
 README_FILE = "README.md"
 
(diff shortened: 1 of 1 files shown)


alecthomas:
This is deliberate - there's a script in ./scripts

JakobDev:
This should probably better documented

alecthomas:
It's true, I should really have a CONTRIBUTING.md I think... 🤦‍♂️

PR #1205, opened 2026-01-26
Add `.qrc` as extension for XML

JakobDev:
This extension [is used by Qt](https://doc.qt.io/qt-6/resources.html#the-qt-resource-compiler-rcc)


alecthomas:
Thanks!

PR #1206, opened 2026-01-27
Update the PHP lexer

vrana:
This updates the PHP lexer from Pygments. Namely, it adds support for [attributes](https://www.php.net/manual/en/language.attributes.overview.php). I've preserved the `_` handling in numbers which is not available in Pygments and I've also improved it a bit.


alecthomas:
Thanks!

PR #1207, opened 2026-01-27
Fix error handling in `needRules()`

JakobDev:
I discovered a little bug while trying to import a Pygments lexer: `needRules()` calls `fetchRules()` inside `compileOnce`. This means that `fetchRules()` is only called the first time `needRules()` is called. So if the `needRules()` is called by `Tokenise()` the error will only be handled the first time. Any call to `Tokenise()` after that will cause a panic, as no rules are loaded. To fix that, we need to store the error in the struct.

If you want to test it, you can try this invalid lexer in `chromad`:
```xml
<lexer>
  <config>
    <name>Test</name>
  </config>
  <rules>
  </rules>
</lexer>
```
Just select the Lexer and start typing.

### DIFF
--- a/regexp.go
+++ b/regexp.go
@@ -308,6 +308,7 @@ type RegexLexer struct {
 	rules          map[string][]*CompiledRule
 	fetchRulesFunc func() (Rules, error)
 	compileOnce    sync.Once
+	compileError   error
 }
 
 func (r *RegexLexer) String() string {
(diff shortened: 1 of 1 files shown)


alecthomas:
Nice catch, thanks. I'm actually pretty surprised the linter didn't pick up that `err` not being used...

JakobDev:
> I'm actually pretty surprised the linter didn't pick up that err not being used...

The problem is, that `err` is only set the first time the function is called. That's hard for an linter to catch, as it's a logic error.

alecthomas:
[errcheck](https://github.com/kisielk/errcheck) is designed to pick up exactly this situation: error values written to but not read from.

PR #1208, opened 2026-01-27
Show error message in chromad

JakobDev:
This will show any error that is thrown by chroma in the web interface of chromad


alecthomas:
I think ideally this would go in a separate error alter, but this is okay.

Issue #1062, opened 2025-03-17
Support a real Luau lexer


CompeyDev:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What is the missing lexer?

Currently, there is no Luau lexer, and Luau files are simply aliased to the Lua lexer as per #1056. The correct implementation would be to have a real Luau lexer for the Luau language, which has significantly diverged from Lua. 

Such an alias tends to often misbehave for syntax highlighting due to how similar tokens can have entirely different meanings in contexts -- and has been evident in the past when users used to use `.gitattributes` overrides to alias Luau to Lua before official support was landed in GitHub's Linguist library.

### Links to existing syntax definitions

There is a community maintained repository containing Textmate language files, which the Linguist library also uses: [JohnnyMorganz/Luau.tmLanguage](https://github.com/JohnnyMorganz/Luau.tmLanguage).

PR #1209, opened 2026-01-30
Import Luau lexer

JakobDev:
This imports `pygments.lexers.scripting.LuauLexer` from Pygments.

Some notes:
- The generated XML file is not 100% valid as as some elements have 2 `state` attributes, but it does work
- I needed to tweak the import script to correctly convert the RegEx
- I needed to add `OperatorReserved` as a `TokenType`
- `.luaurc` is added as a filename to the JSON Lexer, as [it's used by Luau](https://rfcs.luau.org/config-luaurc.html)

Implements #1062


JakobDev:
> Would it make more sense to integrate this into the Lua lexer?

Luau is based on Lua but it's a separate Language. See [the documentation](https://luau.org/compatibility) for more details.

Also Pygments has this as a separate lexer. It would be to much work to integrate this into the existing Lua lexer.

JakobDev:
Ping @alecthomas

PR #1219, opened 2026-02-21
Add AMPL lexer

nunnsy:
Add support for [AMPL](https://ampl.com/).

Largely based on [`pygments.lexers.ampl.AmplLexer`](https://github.com/pygments/pygments/blob/master/pygments/lexers/ampl.py) from Pygments.


nunnsy:
Happy to manually squash these sorry - had a few passes at this given the Pygments implementation isn't 100% true to the syntax.

lexers/embedded/ampl.xml
```diff
@@ -0,0 +1,98 @@
+<lexer>
+  <config>
+    <name>AMPL</name>
+    <alias>ampl</alias>
+    <filename>*.mod</filename>
+    <filename>*.run</filename>
```
alecthomas: What unfortunate file extensions...very common

PR #1220, opened 2026-02-22
CONTRIBUTING.md & Fixing README

nunnsy:
- Adds a `CONTRIBUTING.md` document, as I found some of the instructions sparse. I realise this is likely to require edits from @alecthomas given it's his project.
- Updates the `table.py` script to actually dynamically update the `README.md` file.
- Regenerate the `README.md` with the updated above script.

**Side note**, is it also possible to create a release if/when this PR is merged so the new lexers are available downstream in Hugo? Tyvm

### DIFF
--- a/table.py
+++ b/table.py
@@ -18,6 +18,7 @@ lexers:
     aliases: abnf
     filenames: *.abnf
     mimetypes: text/x-abnf
+...
 
 """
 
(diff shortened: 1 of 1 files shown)


nunnsy:
Ah, I've just seen https://github.com/alecthomas/chroma/pull/1203/ - not sure this is a desired change off the back of that. cc @JakobDev

PR #1228, opened 2026-03-13
Add ERB lexer

nelsonfigueroa:
This PR adds support for embedded Ruby ([ERB](https://github.com/ruby/erb)) which is commonly used in Ruby on Rails projects. Note that I used heavy assistance from Claude Code and tried to ensure it didn't generate slop to the best of my abilities. All tests are passing and I also visually verified the end result which looks solid to me.

Here's a screenshot that was generated by building the Chroma CLI with the ERB lexer and running it against the test data file with `chroma --lexer=erb --style=monokai --html lexers/testdata/erb.actual`

image


alecthomas:
Thanks!

PR #1244, opened 2026-04-16
feat(lexers): add support for Flint language

lucaas-d3v:
This PR adds a lexer for the [Flint](https://github.com/the-flint-lang/flint) programming language.

Flint is a fast, pipeline-oriented AOT language for system scripting that compiles to C99.

**Changes:**
- Added `lexers/flint.go` covering keywords, built-ins, types, and the pipeline operator (`~>`).
- Tested with `go test ./lexers/...` — passed.

### DIFF
--- a/lexers/flint.go
+++ b/lexers/flint.go
@@ -1,0 +1,54 @@
+package lexers
+
+import (
+	"github.com/alecthomas/chroma/v2"
+)
+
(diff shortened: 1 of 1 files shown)


alecthomas:
Please convert this to an XML lexer, Go lexers are used only under specific circumstances.

PR #1268, opened 2026-05-17
Add YAML+Jinja lexer

ktrysmt:
## Summary

Adds a `YAML+Jinja` lexer that combines the existing YAML and Django/Jinja lexers via `DelegatingLexer`. This is the equivalent of Pygments' [`YamlJinjaLexer`](https://pygments.org/docs/lexers/#pygments.lexers.templates.YamlJinjaLexer) and covers two common formats that the current lexers cannot highlight correctly on their own:

- Ansible playbooks (Jinja2 expressions embedded in YAML)
- Salt SLS files

Implementation follows the same pattern as `ERB`, `PHTML`, `Svelte`, and `Go HTML Template`: a small Go file wraps two existing lexers with `DelegatingLexer`, reusing `embedded/yaml.xml` and `embedded/django_jinja.xml`. No new XML file is introduced.

## Lexer config

| Field | Value |
| ----- | ----- |
| Name | `YAML+Jinja` |
| Aliases | `yaml+jinja`, `salt`, `sls`, `ansible` |
| Filenames | `*.sls` |
| MIME types | `text/x-yaml+jinja`, `text/x-sls` |

### Why no `*.yml` / `*.yaml`?

Ansible playbooks and Salt SLS files often use `*.yml`. Claiming those globs would route every plain YAML file through this lexer and create an ambiguous match against the built-in `YAML` lexer. To stay non-invasive, the lexer only claims `*.sls`; for Ansible playbooks the user is expected to select the lexer explicitly (`chroma -l ansible …`), matching Pygments' behavior.

Existing lexers remain unaffected:

| Input | Resolved lexer |
| ----- | -------------- |
| `*.yml` / `*.yaml` | `YAML` (unchanged) |
| `*.sls` | `YAML+Jinja` (new) |
| `text/x-yaml` | `YAML` (unchanged) |
| `application/x-jinja` | `Django/Jinja` (unchanged) |
| `text/x-yaml+jinja`, `text/x-sls` | `YAML+Jinja` (new) |

## Test plan

- [x] `go test ./...` passes (new `TestLexers/YAML+Jinja/testdata/ansible.actual` case included; `.expected` generated via `RECORD=true go test ./lexers`)
- [x] `golangci-lint run ./...` passes
- [x] `gofmt -l` clean
- [x] Manual smoke test: `lexers.Get("ansible")`, `lexers.Get("yaml+jinja")`, `lexers.Get("salt")`, `lexers.Get("sls")` all resolve to the new `YAML+Jinja` lexer
- [x] Verified that `*.yml`, `*.yaml`, `text/x-yaml`, and `application/x-jinja` still resolve to the previous lexers (no regression)


alecthomas (approved):

alecthomas:
Have you verified that it renders how you'd expect?

ktrysmt:
Yes — I rendered lexers/testdata/ansible.actual with the new lexer at Monokai and the HTML formatter, and verified visually:
 
 - YAML structure (keys / scalars / booleans / numbers / list markers) is colored as plain YAML would be
 - Jinja delimiters ({{ }}, {% %}, {# #}) are tokenized as CommentPreproc / Comment and clearly separated from the surrounding YAML
 - Jinja keywords (if, endif), variables (ansible_user, package_name, inventory_hostname, …), filter pipe |, function call default(...), and inner string literals ('root') are all highlighted distinctly
 - Both quoted ("{{ x }}", "files/{{ x }}.pem") and unquoted (Ensure {{ x }} is installed, /etc/ssl/certs/{{ x }}.pem) embeddings work as expected

screenshot

alecthomas:
Thanks!

PR #1278, opened 2026-05-30
Fix YAML slash comments with colons

### LINKED ISSUE #1271: Fix highlight for YAML inline comments with a colon

- [x] I have searched the existing issues

### Describe the bug

YAML highligting breaks on inline comments using colon.

Example:
Image

Expectation: After "//" all text should be treated as an inlined comment

It affects all themes

### To Reproduce

Go to : https://swapoff.org/chroma/playground/
...and pate:
```yaml
foo: "Foo" // default
bar: "Bar" // default: Bar
```

cyphercodes:
Fixes #1271.

## Summary
- Treat `//` inline comment text as `Comment` in YAML highlighting.
- Keep key matching from running through `//` comment text before a later colon.
- Add YAML lexer fixture coverage for `//` comments with colon text.

## Tests
- `PATH=/root/sdk/go1.26.2/bin:$PATH go test ./lexers -run 'TestLexers/YAML' -count=1`
- `PATH=/root/sdk/go1.26.2/bin:$PATH go test ./... -count=1`
- `git diff --check`


alecthomas (approved):

alecthomas:
Thanks!

PR #1277, opened 2026-05-28
Add templ lexer

bryce-d-greybeard:
## Summary

Adds a lexer for [templ](https://templ.guide/), the Go HTML component/template language used by `*.templ` files.

The lexer handles templ declarations, Go-like control/code expressions, component calls such as `@Button(...)`, and delegates ordinary HTML markup/text to the existing HTML lexer.

## Tests

- `go test ./lexers`
- `go test ./...`

PR #1283, opened 2026-06-06
Add templ XML lexer

bryce-d-greybeard:
Adds a templ lexer as an XML lexer definition under `lexers/embedded/templ.xml`.

This replaces the earlier Go-based lexer approach from #1277, following maintainer feedback to use an XML lexer instead.

Includes fixtures for:
- templ component declarations
- HTML tags and attributes
- Go expressions in `{ ... }`
- braced attribute values
- templ control flow (`if` / `else`)
- component calls with `@Button(...)`
- analyser coverage

Local test gates:
- `go test ./lexers -count=1`
- `go test ./... -count=1`


alecthomas:
Thanks!

PR #1294, opened 2026-06-27
Add missing NameAttribute entry to github-dark.xml

bep:
* NameAttribute has the same color as NameClass in github.xml, so I followed that same pattern.
* There are some other missing types in github-dark.xml, but this was the most obvious missing piece looking at some samples.


alecthomas (approved):

alecthomas:
Thanks!

PR #1329, opened 2026-07-26
feat(lexers): add Plan 9 rc lexer

enzv:
Go has always had a good relationship with Plan 9. That is still visible today. Plan 9 remains a supported Go target, and rc is the shell at the center of that world, roughly the Plan 9 counterpart to bash but with a much more Plan 9-shaped syntax.

This PR adds a Plan 9 rc lexer with custom Go handling for context-sensitive forms that do not fit well in XML alone, including heredocs, command and process substitution, decorated redirections, pipes, quoting, and rc-specific control syntax.

https://github.com/plan9foundation/plan9/blob/main/sys/man/1/rc


alecthomas (approved):

alecthomas:
Thanks!

PR #1332, opened 2026-07-29
fix(docker): tokenize newlines as whitespace instead of Error

hdimer:
## Summary

The Dockerfile lexer emits an `Error` token for the trailing newline of every instruction, even for a minimal `FROM alpine`:

```
FROM alpine
```
tokenises as `Keyword("FROM")`, `TextWhitespace(" ")`, `LiteralString("alpine")`, `Error("\n")`. A realistic multi-instruction Dockerfile produces one `Error` token per line.

## Root cause

The `root` state has no rule matching a bare newline, so once an instruction rule stops at the end of a line, the `\n` falls through to no rule and becomes an `Error`.

This lexer is ported from Pygments, whose `RegexLexer` has an implicit fallback: when no rule matches and the current character is `\n`, it emits `Text` and resets to `root` (see `get_tokens_unprocessed`). Chroma has no such implicit newline handling, so a Pygments-ported lexer that relied on it needs the rule spelled out.

## Fix

Add a whitespace rule to the `root` state:

```xml
<rule pattern="\s+">
  <token type="TextWhitespace" />
</rule>
```

Newlines (and blank lines) now tokenise as `TextWhitespace`; `FROM ... AS`, `\`-continuations, `HEALTHCHECK`, `CMD [...]`, etc. are unaffected, and there are no more `Error` tokens.

## Tests

There was no Docker test fixture, so this adds one (`lexers/testdata/docker.actual` + recorded `docker.expected`) covering `FROM ... AS`, `WORKDIR`, `COPY`, `ENV`, `ARG`, a line-continuation `RUN`, `EXPOSE`, `HEALTHCHECK`, and a JSON-form `CMD`. `go test ./...` passes.

Note: the same "unmatched newline becomes Error" pattern likely affects other Pygments-ported lexers; this PR fixes only Docker, happy to look at a broader sweep separately if that's wanted.

---

Disclosure: this fix was developed with AI assistance; I reviewed it and stand behind it.


alecthomas (approved):

alecthomas:
Thanks!

PR #1344, opened 2026-08-12
Add support for modern IRC formatting

### LINKED ISSUE #900: Add mIRC formatter support

### What problem does this feature solve?

I'm currently using chroma in another project, [matterircd](https://github.com/42wim/matterircd). It is a [Mattermost](https://mattermost.com/) & Slack chat to IRC bridge. chroma is used for [converting code blocks with syntax highlighting](https://docs.mattermost.com/collaborate/format-messages.html#code-blocks) to pretty output.

The current use is using `terminal256` by default. This works fine for console based IRC clients with support for terminal colors but it does not for others without support such as [WeeChat](https://weechat.org/) and non-terminal based IRC clients.

### What feature do you propose?

Modern IRC clients support hex colors - https://modern.ircdocs.horse/formatting.html#hex-color

It would be nice if chroma could add another formatter such as `mirc256` with the generated output using IRC's ASCII 0x04 plus the hex colors.

hloeung:
Fixes #900 adding formatters `mirc100` for extended color codes and `mirc16m` for RGB / hex colors.

Started with a copy of https://github.com/alecthomas/chroma/blob/v2/formatters/tty_truecolour.go because it is very similar.

For extended IRC colors (0-99), that's https://modern.ircdocs.horse/formatting.html#colors-16-98. For this, we have to build a palette of colors and their RGB hex color codes. Then from a given RGB hex color code, find the closest to use in our map. It also supports background colors with the addition of appending `,<CODE>`.

For hex color codes, that's https://modern.ircdocs.horse/formatting.html#hex-color. We can use straight `entry.Colour.Red(), entry.Colour.Green(), entry.Colour.Blue()`. Unfortunately, it doesn't look like it supports background colors so I have left that out.

Tested and seems to be working great, see https://github.com/42wim/matterircd/pull/783.

### DIFF
--- a/formatters/mirc.go
+++ b/formatters/mirc.go
@@ -1,0 +1,209 @@
+package formatters
+
+import (
+	"fmt"
+	"io"
+	"math"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/alecthomas/chroma/v2"
+)
+
+// mIRC extended color codes
+var mIRC100 = Register("mirc100", chroma.FormatterFunc(mIRCExtendedColorFormatter))
+
+// mIRC16m is a true-colour formatter.
+var mIRC16m = Register("mirc16m", chroma.FormatterFunc(mIRCHexColorFormatter))
+
+// Print the text with the given formatting, resetting the formatting at the end
+// of each line and resuming it on the next line.
+//
+// This way, a pager (like https://github.com/walles/moar for example) can show
+// any line in the output by itself, and it will get the right formatting.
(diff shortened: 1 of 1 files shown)


alecthomas:
Why are there two pull requests?

hloeung:
> Why are there two pull requests?

This one is targetted to v2, different set of changes... slightly.

alecthomas:
Ah! I'm no longer accepting PRs against v2. Thanks anyway

hloeung:
Ah okay, thanks though.

PR #1346, opened 2026-08-15
fix(svg): non-ASCII text misaligns token background rects

VXNCXNX:
The SVG formatter measures line widths in bytes rather than characters, so any non-ASCII rune inflates the canvas and pushes token background rects out of alignment with the text they sit behind.

Tokenising `x = ["éé", "ab"]` with a style that sets a string background:

```
before: <svg width="136px" ...>
        <rect id="&quot;éé&quot;" x="4ch"  ... width="6ch" />
        <rect id="&quot;ab&quot;" x="12ch" ... width="4ch" />

after:  <svg width="120px" ...>
        <rect id="&quot;éé&quot;" x="4ch"  ... width="4ch" />
        <rect id="&quot;ab&quot;" x="10ch" ... width="4ch" />
```

`"éé"` is 4 characters and 6 bytes. The old code gives its rect a width of 6 cells, then starts the next rect 2 cells too far right, so the `"ab"` highlight is visibly offset from `"ab"`. Every token after the first non-ASCII one on a line inherits that drift, and the `<svg width>` is overstated by the same amount.

`x` and `width` are emitted in `ch` units, which are character cells in a monospace font, so the count has to be in runes.

## The fix

`maxLineWidth` and `writeTokenBackgrounds` computed the same expression with `len()`. Extracted it as `tokenWidth` using `utf8.RuneCountInString`, keeping the existing tab expansion.

Pure-ASCII output is unchanged. Rendering `lexer.go` from this repo to SVG before and after gives byte-identical files.

## Verification

`TestNonASCIIWidths` in `formatters/svg/svg_test.go` builds a style with a string background so the rect path is exercised, then asserts the canvas width and both rect offsets.

Reverting `svg.go` fails it, and the failure output shows the misalignment directly:

```
Needle:   <svg width="120px"
Haystack: ... <svg width="136px" ...
          <rect id="&quot;éé&quot;" x="4ch" ... width="6ch" ... />
          <rect id="&quot;ab&quot;" x="12ch" ... width="4ch" ... />
```

`go test ./...` passes and `golangci-lint run` is clean.

*Disclosure: written with AI assistance (Claude Code). I produced the before and after by rendering real files through both builds, and ran the revert check myself.*

### DIFF
--- a/formatters/svg/svg.go
+++ b/formatters/svg/svg.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"slices"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/alecthomas/chroma/v3"
 )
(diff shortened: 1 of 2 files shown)


alecthomas:
@codex review

PR #1348, opened 2026-08-15
fix(cmd/chroma): --html-styles ignores the other --html-* flags

VXNCXNX:
Stacked on #1347, which is needed for `cmd/chroma` to build at all.

`--html-styles` ignores every other `--html-*` flag.

```
$ chroma --html-styles --html-prefix=foo- --html-tab-width=4 | head -2
/* Background */ .bg { color: #e5e5e5; background-color: #000000; }
/* PreWrapper */ .chroma { color: #e5e5e5; background-color: #000000; -webkit-text-size-adjust: none; }
```

After:

```
/* Background */ .foo-bg { color: #e5e5e5; background-color: #000000;tab-size: 4; }
/* PreWrapper */ .foo-chroma { color: #e5e5e5; background-color: #000000;tab-size: 4; -webkit-text-size-adjust: none; }
```

Generating a stylesheet is exactly when `--html-prefix` matters most, since the prefix has to agree with whatever generated the HTML. Emitting unprefixed CSS produces a stylesheet that silently matches nothing.

## Cause

```go
if cli.Formatter == "html" {
	configureHTMLFormatter(ctx)
}
```

but `--html-styles` writes CSS from the html formatter regardless of which formatter is selected. So unless you also pass `-f html`, which is pointless when you only want the stylesheet, the html options are never applied.

Same for `--html-lines-table`, `--html-highlight`, `--html-linkable-lines` and the rest of the group.

## The fix

A small `usesHTMLFormatter()` covering both cases, at the one call site.

Unchanged, checked before and after: `-f html --html-prefix=zz-` still emits `zz-chroma`, and a bare `--html-styles` with no other flags is byte-identical.

## Verification

`TestHTMLStylesAppliesHTMLOptions` in `cmd/chroma/main_test.go` sets `cli.Formatter = "terminal"` with `--html-styles`, `--html-prefix` and `--html-tab-width`, then asserts both appear in the CSS.

Reverting the condition to `cli.Formatter == "html"` fails it:

```
--- FAIL: TestHTMLStylesAppliesHTMLOptions
    main_test.go:29: expected html formatter to be in use
```

`go test ./...` passes in both the root module and `cmd/chroma`, and `golangci-lint run` at the root is clean.

*Disclosure: written with AI assistance (Claude Code). I built both binaries, produced the before and after above by running them, and ran the revert check myself.*

### DIFF
--- a/cmd/chroma/go.mod
+++ b/cmd/chroma/go.mod
@@ -12,6 +12,6 @@ require (
 )
 
 require (
-	github.com/dlclark/regexp2/v2 v2.5.2 // indirect
+	github.com/dlclark/regexp2/v2 v2.7.1 // indirect
 	golang.org/x/sys v0.29.0 // indirect
 )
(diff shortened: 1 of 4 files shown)


alecthomas:
@codex review

PR #1349, opened 2026-08-16
fix(lexers): the Markless supertext rule was unreachable

VXNCXNX:
The Markless `inline` state includes `subtext` twice, so the `supertext` rule right below it is defined but never reachable. Superscript markup comes out as plain text.

```go
Include("subtext"),
Include("subtext"),   // <- this should be supertext
```

Tokenising `Water is H v(2) O and E = mc ^(2) here.`:

```
before:  Text     "Water is H "
         Keyword  "v("
         Text     "2"
         Keyword  ")"
         Text     " O and E = mc ^(2) here."

after:   Text     "Water is H "
         Keyword  "v("
         Text     "2"
         Keyword  ")"
         Text     " O and E = mc "
         Keyword  "^("
         Text     "2"
         Keyword  ")"
         Text     " here."
```

Subscript already worked, which is why it went unnoticed: the duplicated include still made `v(...)` highlight, so only the superscript half was silently dead.

## The fix

One line. The rule itself, `{`(\^\()(.*?)(\))`, ByGroups(...)}`, is unchanged and was already written correctly; it just had no way in.

## Verification

A superscript case added to `lexers/testdata/markless.actual`, with `markless.expected` regenerated via `RECORD=true go test ./lexers/`, which is the mechanism documented in `lexers_test.go`. The diff to the expected file is append-only, so nothing else re-tokenised.

Reverting only line 105 back to `Include("subtext")` fails it:

```
--- FAIL: TestLexers/Markless/testdata/markless.actual
    -  {"type":"Text","value":" O, and the area is 4 "},
    -  {"type":"Keyword","value":"^("},
    -  {"type":"Text","value":"2"},
    -  {"type":"Keyword","value":")"},
    +  {"type":"Text","value":" O, and the area is 4 ^(2) cm."},
```

`go build ./...`, `go test ./...`, `gofmt -l` and `golangci-lint run ./lexers/` are all clean.

*Disclosure: written with AI assistance (Claude Code). I ran the tokeniser before and after, regenerated the fixture with the documented RECORD mechanism, and ran the mutation check myself.*

### DIFF
--- a/lexers/markless.go
+++ b/lexers/markless.go
@@ -102,7 +102,7 @@ func marklessRules() Rules {
 			Include("compound"),
 			Include("footnote-reference"),
 			Include("subtext"),
-			Include("subtext"),
+			Include("supertext"),
 			Include("url"),
 			{`.`, Text, nil},
 			{`\n`, TextWhitespace, Pop(1)},
--- a/lexers/testdata/markless.actual
+++ b/lexers/testdata/markless.actual
@@ -43,3 +43,5 @@ The Markless document standard and ecosystem is brought to you by the ''Shirakum
 
 1. First things first, this <-is great->
 10. Now for something different.
+
+The formula is H v(2) O, and the area is 4 ^(2) cm.
--- a/lexers/testdata/markless.expected
+++ b/lexers/testdata/markless.expected
@@ -161,5 +161,15 @@
   {"type":"Keyword","value":"10."},
   {"type":"TextWhitespace","value":" "},
   {"type":"Text","value":"Now for something different."},
+  {"type":"TextWhitespace","value":"\n\n"},
+  {"type":"Text","value":"The formula is H "},
+  {"type":"Keyword","value":"v("},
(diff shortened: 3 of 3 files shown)


alecthomas:
@codex review

VXNCXNX:
Bump on this one, and on the two beside it, since you ran `@codex review` across all three on 24 Aug and it came back clean on each.

All three are still green and still merge without conflict:

| | Lint | Test | state |
| --- | --- | --- | --- |
| #1349 the Markless supertext rule was unreachable | pass | pass | clean |
| #1352 Svelte 5 snippet blocks lex as errors | pass | pass | clean |
| #1353 markdown code spans and strikethrough match across a blank line | pass | pass | clean |

Each touches one lexer file plus its `.actual`/`.expected` pair and nothing else, so they are independent and can go in in any order, or separately if only some of them appeal.

Anything you want changed on any of them, or is there something else they need? Happy to close any that are not wanted.

PR #1352, opened 2026-08-17
fix(lexers): Svelte 5 snippet blocks lex as errors

VXNCXNX:
## What's broken

The Svelte lexer emits `Error` tokens for Svelte 5 `{#snippet}` blocks, so they render as errors. Reported in #1306.

## Repro

```
$ echo '{#snippet figure(caption)}<figure>{caption}</figure>{/snippet}' \
    | chroma --lexer svelte --formatter tokens

before:
  Token{Punctuation, "{"}
  Token{Error, "#"}
  Token{NameOther, "snippet"}
  ...
  Token{Error, "/snippet"}

after:
  Token{Punctuation, "{"}
  Token{Keyword, "#snippet"}
  ...
  Token{Keyword, "/snippet"}
```

## The fix

The block-keyword rule listed `await|each|if|key`, which are the Svelte 4 blocks. `snippet` was added in Svelte 5. One word added to the alternation.

I checked the rest of the Svelte 5 surface rather than only the reported case: `{@render}`, `{@const}`, `{@html}`, `{@debug}`, `{#await}`, `{#each}`, `{#if}` and `{#key}` all already lex with no `Error` token, so `snippet` was the only gap and this needs no follow-up.

## Verification

Extended `testdata/svelte.actual` with a snippet block and regenerated `.expected`. Removing `snippet` from the rule fails `TestLexers/Svelte` with the `Error` tokens shown above, not a build error. `go test ./...` passes.


alecthomas:
@codex review

PR #1353, opened 2026-08-17
fix(lexers): markdown code spans and strikethrough match across a blank line

VXNCXNX:
## What's broken

An unclosed backtick span swallows everything up to the next backtick anywhere in the document, including headers, which lose their highlighting. Reported in #1325.

## Repro

```
$ chroma --lexer markdown --formatter tokens
here is `an open span

## Swallowed Header

and it closes`

before:
  Token{Text, "here is "}
  Token{LiteralStringBacktick, "`an open span\n\n## Swallowed Header\n\nand it closes`"}

after:
  Token{Text, "here is `an open span\n\n"}
  Token{GenericSubheading, "## Swallowed Header\n"}
```

## The fix

The rule was `` `[^`]+` ``, and the negated class matches newlines, so the span could run through blank lines. A blank line ends the paragraph, so a code span cannot cross one (CommonMark 6.1). The rule now allows a line ending but not a blank one.

`~~strikethrough~~` had the same root cause, so it is fixed in the same way. `*emph*` and `**strong**` already scope correctly, so I left them alone.

## Verification

Two cases added to `testdata/markdown.actual`, one per rule. Reverting each rule separately fails its own case with the swallowed header reappearing as one token.

I checked the fix does not over-correct: spans and strikethroughs that wrap a single line break still lex as before, and I diffed the token output against a binary built from master to confirm those cases are byte-identical.

`go test ./...` passes across all packages.

### DIFF
--- a/lexers/markdown.go
+++ b/lexers/markdown.go
@@ -98,8 +98,10 @@ func markdownRules() Rules {
 			{`\\.`, Text, nil},
 			{`(\s)(\*|_)((?:(?!\2).)*)(\2)((?=\W|\n))`, ByGroups(Text, GenericEmph, GenericEmph, GenericEmph, Text), nil},
 			{`(\s)((\*\*|__).*?)\3((?=\W|\n))`, ByGroups(Text, GenericStrong, GenericStrong, Text), nil},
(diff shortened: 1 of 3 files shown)


alecthomas:
@codex review

PR #1354, opened 2026-08-18
perf(styles): parse embedded styles lazily

mattn:
Importing the styles package parses all 79 embedded style XML files at init. This parses each style lazily on first lookup instead, keyed by file name, which by convention is the lowercased style name — a new test enforces that invariant and that every embedded style parses. Only the swapoff fallback is parsed eagerly, and the registry is now mutex-guarded, so concurrent use is safer than before.

|---|---|---|
| before | 5.4 ms | 2.2 MB / 44,621 allocs |
| after | 0.06 ms | 14 KB / 225 allocs |

### DIFF
--- a/styles/api.go
+++ b/styles/api.go
@@ -16,7 +16,7 @@ var embedded embed.FS
 var (
 	registryMu sync.Mutex
 	// registry holds user-registered styles plus embedded styles that have
-	// already been parsed. Embedded styles are parsed lazily by lookup so
+	// already been parsed. Embedded styles are parsed lazily by Lookup so
 	// that importing this package does not pay for parsing every style;
 	// see TestEmbeddedStyleNamesMatchFilenames for the invariant that makes
 	// the lazy filename-based lookup possible.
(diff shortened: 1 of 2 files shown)


styles/api.go
```diff
@@ -3,67 +3,98 @@ package styles
...
+		names = append(names, strings.TrimSuffix(file.Name(), ".xml"))
 	}
-	return r
-}()
+	return names
+})
+
+// lookup returns the style with the given lowercased name, parsing it from
+// the embedded styles on first use. registryMu must be held.
+func lookup(name string) (*chroma.Style, bool) {
```
alecthomas: Doesnt this need to take the lock?

It's called from two locations - one of which takes the lock, but the other doesn't. Might be simpler to just inline this into `Lookup()` along with the lock.

mattn: Good catch — inlined `lookup` into `Lookup()` so the parse always happens under the lock, and `Fallback` now goes through `Lookup()` as well. Fixed in 20fbc3c.

alecthomas (approved):

PR #1355, opened 2026-08-18
perf(svg): defer embedded font encoding until output is written

mattn:
Registering the default SVG formatter base64-encodes the embedded 390 KB Liberation Mono font at init, in every program importing the formatters package. This adds an EmbedFontProvider option that takes the font as a func() string and defers the call until the formatter actually writes output. EmbedFont is unchanged and now wraps it.

|---|---|---|
| before | 0.57 ms | 509 KB |
| after | 0.14 ms | 34 KB |


alecthomas (approved):
Thanks!

PR #1356, opened 2026-08-18
perf(lexers): build the registry from generated metadata instead of parsing XML at init

mattn:
Importing the lexers package decodes the `<config>` element of all 285 embedded XML definitions with encoding/xml at init. This generates that metadata into Go literals with a new go:generate tool (internal/metadatagen, output printed with repr), so building the registry parses no XML at all; each lexer's rules are still loaded lazily from its XML on first use, as before. A test verifies the generated metadata matches the XML and fails with a hint to re-run go generate when they drift.

Analyser regexes are now compiled on the first AnalyseText call, with their validity checked at generation time instead. Registering a fresh lexer also no longer scans the whole registry slice, which was quadratic over a full registration pass.

|---|---|---|
| before | 9.4 ms | 3.0 MB / 34,714 allocs |
| after | 0.92 ms | 463 KB / 2,341 allocs |


lexers/lexers_gen.go
```diff
@@ -0,0 +1,4132 @@
...
+package lexers
+
+import (
+	"github.com/alecthomas/chroma/v3"
+)
+
+// embeddedLexers mirrors the <config> element of every XML lexer definition
+// under embedded/, so that the registry can be built without parsing XML at
+// init time. TestEmbeddedLexerMetadata verifies it is up to date.
+var embeddedLexers = []struct {
```
alecthomas, line 12: How much bigger does this make the compiled code? One of the primary reasons for switching to XML was to reduce binary size.

internal/metadatagen/main.go
```diff
@@ -0,0 +1,85 @@
+// Command metadatagen generates a Go source file containing the <config>
+// section of every embedded lexer XML definition, so that the lexer registry
+// can be constructed without parsing any XML at init time.
+package main
```
alecthomas, line 4: If the compiled size doesn't end up being too bad, this will need a new `just` command to cross-check that the generated code is in sync with the builtins, and that will need to be wired up to CI.

lexers/lexers_gen.go
mattn: This turned out smaller than I expected. Measured with a minimal program importing `lexers`, master vs this branch (go1.26.1):

| build | before | after | delta |
|---|---|---|---|
| linux/amd64 | 7,848,296 | 7,855,033 | +6.7 KB (+0.09%) |
| linux/amd64 `-ldflags=-s -w` | 6,189,218 | 6,221,986 | +32.8 KB (+0.53%) |
| `cmd/libchromawasm` (GOOS=js GOARCH=wasm) | 9,601,933 | 9,639,391 | +37.5 KB (+0.39%) |

The growth stays small because only the `<config>` metadata (names, aliases, globs, MIME types) is generated into Go code — the lexer rules, which dominate the size, remain in the embedded XML and are still loaded lazily. The metadata is duplicated between the XML and the generated code, but the duplication compiles to ~30 KB.

internal/metadatagen/main.go
mattn: Added in c159415: a `just check-generated` recipe that re-runs `go generate ./lexers` and fails with a hint if `lexers/lexers_gen.go` differs from what is committed, plus a CI job running it. (The metadata is also cross-checked against the XML at test time by `TestEmbeddedLexerMetadata`, so `go test` catches drift too.)

alecthomas: 🙏

lexers/lexers_gen.go
alecthomas: Niceeee, that's very tolerable, thanks for checking.

alecthomas (approved):

Issue #722, opened 2023-01-10
With LineNumbersInTable enabled, highlighted line numbers have wrong height


jmooring:
With `LineNumbersInTable` enabled, if the inherited CSS line height is greater than one, then the height of highlighted numbers is not equal to the height of the highlighted code.

![image](https://user-images.githubusercontent.com/335669/211472836-ecafff3c-8568-405c-8031-2e464b77f15f.png)

@willfaught suggested the following fix:

```css
.chroma .lntable .lnt,
.chroma .lntable .hl {
    display: flex;
}
```

And while the body is open, perhaps we could make the second column take up the remaining width. You can see the difference above.

Using the [adjacent sibling combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator):

```css
.lntable .lntd + .lntd  {
  width: 100%;
}
```

With both of these changes:

![image](https://user-images.githubusercontent.com/335669/211476193-f15a0f28-5f06-4197-8f75-9992da2bd2b1.png)

jmooring:
Please re-open. The problem is still present in v2.23.0. Thanks.

And for anyone stumbling across this...  yes, you can work around the problem with inline line numbers, but then you run into this 9 year old Firefox bug: <https://bugzilla.mozilla.org/show_bug.cgi?id=1273836>. However, the Firefox market share is < 3% these days, so maybe that's OK.

alecthomas:
This does seem like a real bug, but there's also not been a PR for the fix in three years even though there seems to be a proposed solution...

jmooring:
Please re-open and I'll put it on my list. If it's still open 6 months from now, close it again.

alecthomas:
Sounds good!

PR #1357, opened 2026-08-18
fix(html): align table line numbers with flex layout

ChrisJr404:
Fixes the long-standing misalignment where, with `LineNumbersInTable` enabled and an inherited `line-height` greater than one, highlighted line numbers are a different height to the highlighted code beside them (#722).

The code column already lays each line out as a flex container (`.chroma .line { display: flex; }`), but the line number column's spans are left inline. When the line-height is stretched, the two columns compute row heights differently, so a highlighted number no longer matches the height of the code on the same row.

This emits a scoped rule from `WriteCSS`, alongside the existing table special-cases, that gives the line number column the same flex layout:

```css
.chroma .lntable .lnt, .chroma .lntable .hl { display: flex; }
```

This is the fix @willfaught proposed on the issue; the companion "second column takes the remaining width" suggestion is already handled by the existing `.lntd:last-child { width: 100% }` rule. The rule is only written when line numbers are actually rendered in a table (`WithLineNumbers` + `LineNumbersInTable`), so default output and the non-table line number modes are unchanged.

Testing: added `TestTableLineNumbersFlex`, which asserts the rule is present for the table mode and absent for plain line numbers. `go test ./...` passes, and `go vet` / `gofmt` are clean.

### DIFF
--- a/formatters/html/html.go
+++ b/formatters/html/html.go
@@ -515,6 +515,21 @@ func (f *Formatter) WriteCSS(w io.Writer, style *chroma.Style) error {
 			return err
 		}
 	}
+	// Special-case the line number column of the table so its lines use the
+	// same flex layout as the code column. The code column's Line already sets
+	// "display: flex", so without this an inherited line-height greater than one
+	// makes the highlighted line numbers a different height to the highlighted
+	// code they sit beside.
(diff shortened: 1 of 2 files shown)


alecthomas:
@codex review

PR #1359, opened 2026-08-25
feat(lexers): add InfluxDB line protocol lexer

sanderson:
Adds a lexer for InfluxDB line protocol, registered under the `lp` and `line-protocol` aliases and the `*.lp` filename.

Elements are tokenised distinctly so themes can style each one: measurement (NameClass), tag keys/values (NameTag/NameLabel), field keys (NameAttribute), field values by type (float, integer/uinteger, string with escapes, boolean), timestamp (LiteralDate), and `#` comments. Invalid content such as unquoted non-numeric field values is emitted as Error tokens.

Also adds it to the supported languages table in the README.


alecthomas (approved):

alecthomas:
Thanks!

PR #1365, opened 2026-08-29
fix(formatters): emit the token background in mirc16m

youdie006:
`mirc16m` is the only colour formatter that drops a token's background.

Four formatters build a formatting prefix from a `chroma.StyleEntry`. Three emit the background;
this one doesn't:

| formatter | file:line | foreground | background |
|---|---|---|---|
| terminal / 8 / 16 / 256 | `formatters/tty_indexed.go:202,205` | `entry.Colour.IsSet()` | `entry.Background.IsSet()` |
| terminal16m | `formatters/tty_truecolour.go:60,63` | `entry.Colour.IsSet()` | `entry.Background.IsSet()` |
| mirc100 | `formatters/mirc.go:172,175` | `entry.Colour.IsSet()` | `entry.Background.IsSet()` |
| **mirc16m** | **`formatters/mirc.go:209`** | `entry.Colour.IsSet()` | **absent** |

`mirc100` is the adjacent function in the same file from the same PR (#1343), so the shape here is
just its twin's.

### Reachable

`clearBackground()` (`formatters/tty_indexed.go:235`) only zeroes the global `chroma.Background`
entry, so per-token backgrounds survive into the formatter. 21 of the 79 bundled styles set one on
`Error` alone — e.g. `styles/monokai.xml:2`, `Error style="#960050 bg:#1e0010"`.

Same token, same style, through each formatter:

```
terminal     "\x1b[35m\x1b[40mbad\x1b[0m"
terminal16m  "\x1b[38;2;150;0;80m\x1b[48;2;30;0;16mbad\x1b[0m"
mirc100      "\x0339,89bad\x0f"
mirc16m      "\x04960050bad\x0f"        <- no background
```

The `,RRGGBB` form is what the hex colour code takes: modern.ircdocs.horse/formatting gives
`0x04RRGGBB,RRGGBB`, matching the `\x03NN,NN` pair `mirc100` already emits.

### Change

Three lines, mirroring `mirc100`. Test added in the file's existing idiom; red before
(`-960050,1E0010WORD` / `+960050WORD`), green after. `go test ./...` passes across all packages,
`gofmt -l` and `go vet` show no change from `master`.

Not included, in case you want it separately: both mIRC formatters nest the background inside
`Colour.IsSet()`, so a background-only entry (`styles/lovelace.xml:2`, `Error style="bg:#a848a8"`)
still drops it in `mirc100` too. That needs the background-only `\x03,NN` / `\x04,RRGGBB` form,
which the spec is less clear about, so I left it alone rather than guess.

---

Disclosure: prepared with AI assistance; I verified the reproduction, the red/green runs and the
test suite myself.

### DIFF
--- a/formatters/mirc.go
+++ b/formatters/mirc.go
@@ -208,6 +208,9 @@ func mIRCHexColorFormatter(w io.Writer, style *chroma.Style, it iter.Seq[chroma.
 		}
 		if entry.Colour.IsSet() {
 			formatting += fmt.Sprintf("\x04%02X%02X%02X", entry.Colour.Red(), entry.Colour.Green(), entry.Colour.Blue())
+			if entry.Background.IsSet() {
+				formatting += fmt.Sprintf(",%02X%02X%02X", entry.Background.Red(), entry.Background.Green(), entry.Background.Blue())
+			}
 		}
 
 		if err := mIRCwriteToken(w, formatting, token.Value); err != nil {
--- a/formatters/mirc_test.go
+++ b/formatters/mirc_test.go
@@ -31,3 +31,18 @@ func TestNoneIRCColour(t *testing.T) {
 
 	assert.Equal(t, "\x0342WORD\x0f", stringBuilder.String())
 }
+
+func TestHexIRCBackgroundColour(t *testing.T) {
+	style, err := chroma.NewStyle("test", chroma.StyleEntries{
+		chroma.Error: "#960050 bg:#1e0010",
+	})
+	assert.NoError(t, err)
+
+	stringBuilder := strings.Builder{}
+	err = MIRC16m.Format(&stringBuilder, style, chroma.Literator(chroma.Token{
+		Type:  chroma.Error,
+		Value: "WORD",
+	}))
+	assert.NoError(t, err)
+	assert.Equal(t, "\x04960050,1E0010WORD\x0f", stringBuilder.String())
+}


hloeung:
Thanks for this - I didn't know that modern IRC also support hex colors for backgrounds but indeed it does.

alecthomas (approved):

youdie006:
Closing the loose end I left in the description, in case anyone picks it up: I said the background-only `\x03,NN` / `\x04,RRGGBB` form was something "the spec is less clear about, so I left it alone rather than guess." I looked it up, and the answer is that there is nothing to do.

modern.ircdocs.horse/formatting enumerates every permitted shape for `0x03`, and background-without-foreground is not one of them:

- `<CODE>` - reset foreground and background
- `<CODE>,` - reset both **and display the `,` as text**
- `<CODE><COLOR>` - set foreground
- `<CODE><COLOR>,<COLOR>` - set both

So emitting `\x03,a8` would reset the colours and then print `,a8` into the highlighted source as literal characters. That is a corruption, not a fix. `0x04` is worse: it has no default-colour sentinel at all (`0x03` has `99`, which the spec itself flags as not universally supported), so `mirc16m` has no expressible background-only encoding even in principle.

The spec also says "If only the foreground color is set, the background color stays the same" - which makes dropping an inexpressible background the faithful rendering rather than a bug.

For the record on size, measured at `4b55271`: 105 background-only entries across 21 of the 79 bundled styles, and both mIRC formatters drop all of them while `terminal16m`/`terminal256` emit them. Real gap, just not one the wire format can close. The only half-measure would be `\x0399,NN` for `mirc100` alone, which leans on the "not universally supported" code and leaves `mirc16m` untouched - recreating the same asymmetry. Not worth it unless you want it.

---

Disclosure: I used Claude (an AI assistant). The counts above are from a probe I ran against this repo, and the spec reading is mine.

PR #1370, opened 2026-09-07
fix(svg): measure token width in columns, not bytes

dylanpulver:
`maxLineWidth` and `writeTokenBackgrounds` size their output with `len()` over a token's UTF-8 **bytes**. Any non-ASCII source therefore gets a canvas measured in bytes, and every background rect after a multi-byte token is pushed right by the extra bytes, so it no longer sits behind the text it highlights.

`<svg width=…>` for a small Python file, monokai:

| source | before | after |
|---|---|---|
| ASCII only | 272px | 272px |
| one line with 3 accented letters | 416px | 392px |
| one 17-character Chinese comment | 464px | 328px |

Counting runes alone would be wrong in the other direction: that Chinese line comes out at 192px, narrower than its glyphs, and the outer `<svg>` clips. So wide and fullwidth runes count as two columns.

Checks: all four new subtests fail on master, and each partial fix (canvas only, rects only, runes without wide handling) is caught by exactly the subtests covering what it left undone. ASCII output is byte-identical before and after. `go test ./...`, `just check-generated` and `golangci-lint run` are green locally.

Known limitation, unchanged here: column counting stays an approximation. Combining marks count as a column each, and the wide table is a fixed set of East Asian blocks, not a live Unicode property.

*Disclosure: written with AI assistance (Claude Code). The before/after numbers come from rendering the same files against each tree, and the mutation checks were run on an unmodified checkout of master.*


alecthomas:
This sounds like a duplicate of #1346

formatters/svg/svg.go
```diff
@@ -141,12 +141,45 @@ func (f *Formatter) writeSVG(w io.Writer, style *chroma.Style, tokens []chroma.T
 	return err
 }
 
+// East Asian Wide and Fullwidth ranges, which a monospace font renders two columns wide.
+var wideRanges = [...][2]rune{
+	{0x1100, 0x115f}, {0x2e80, 0x303e}, {0x3041, 0x33ff}, {0x3400, 0x4dbf},
+	{0x4e00, 0x9fff}, {0xa000, 0xa4cf}, {0xac00, 0xd7a3}, {0xf900, 0xfaff},
+	{0xfe10, 0xfe19}, {0xfe30, 0xfe6f}, {0xff00, 0xff60}, {0xffe0, 0xffe6},
+	{0x1f300, 0x1f64f}, {0x1f900, 0x1f9ff}, {0x20000, 0x3fffd},
+}
```
alecthomas, line 150: This is absolutely not the right way to do this. Go has excellent Unicode support.

PR #1376, opened 2026-09-11
feat(lexers): add TSRX lexer (v2 backport)

austinm911:
Backport of #1375 to the `v2` line. Draft, and it should land after that one (or not at all, if you would rather keep new lexers on `master`).

Same `lexers/embedded/tsrx.xml`, same three cases under `testdata/tsrx/`. `typescript.xml`, which the lexer extends, is byte-identical on `v2` and `master`, and all three `.expected` files recorded on this branch match `master` byte for byte.

Two branch differences:

- no `lexers_gen.go` here, so there is no metadata to regenerate
- `just readme` on `v2` prints the table rather than rewriting `README.md`, so the `TSRX` entry in the `T` row is edited in by hand

`go test ./...` passes.

Why ask at all: consumers still on `v2` cannot pick up a `master`-only lexer without moving to the v3 module path. [ttt](https://github.com/eugenioenko/ttt) pins `chroma/v2 v2.24.1` and is the case that prompted this.


austinm911:
Holding this as a draft until #1375 lands or you say the v2 line is closed to new lexers. The XML and all three recorded token streams are identical to that branch, so this is a rebase away from ready either way.

hloeung:
Maybe @alecthomas can reconsider accepting backports to the v2 line? I have one I'd like to get landed - https://github.com/alecthomas/chroma/pull/1344

alecthomas:
Thanks, but I'm not accepting backports to v2.

PR #1383, opened 2026-09-18
Add Isabelle lexer

Ifiht:
Ported from pygments.lexers.theorem.IsabelleLexer. Covers Isar outer syntax: theory/proof keyword classes, (* *) comments with nesting, cartouches (both ‹...› and \<open>...\<close>, nested), quoted strings and `fact` references with \<symbol> escapes, type variables, and hex/oct/bin literals.

Pygments' nonstandard Name.Type token is mapped to NameDecorator, matching the Standard ML lexer's treatment of 'a type variables.


Ifiht:
This was made (with chagrin)  using Claude Fable 5. It did however pass all tests, the binary built correctly, it was able to lint a real Isabelle theorem from the 2019 Math Olympiad, and passed comparison testing against the AFP's own linted online version of the code:
image
image
https://isa-afp.org/thys/IMO2019/IMO2019_Q1.html

alecthomas (approved):

alecthomas:
Thanks!

PR #1384, opened 2026-09-19
fix(lexers): recognize Python imaginary literals

jakezwang:
Python imaginary literals such as `1j`, `.5j`, `1.2e-3j` and uppercase `J` forms are split into a number and a name. Only exponent forms without a decimal point and with lowercase `j` currently stay together.

Extend the numeric rules to keep the suffix in the `LiteralNumberFloat` token, following the [Python imaginary-literal grammar](https://docs.python.org/3/reference/lexical_analysis.html#imaginary-literals). Add a fixture covering integer, fractional, exponent and underscore forms, with ordinary numeric and identifier controls.

Full tests pass on Go 1.25.13 and with race detection on Go 1.26.6. The pinned golangci-lint 2.12.2 and generated-metadata check also pass.


alecthomas (approved):

alecthomas:
Thanks!

